PR #45, opened 2025-01-24
Fix: constraints in syscall, jump, pc

weilzkm:


crates/core/machine/src/cpu/air/mod.rs
```diff
@@ -50,15 +50,15 @@ where
...
+/*
         // Register constraints.
         self.eval_registers::<AB>(builder, local, is_branch_instruction.clone());
 
         // Memory instructions.
         self.eval_memory_address_and_access::<AB>(builder, local, is_memory_instruction.clone());
         self.eval_memory_load::<AB>(builder, local);
         self.eval_memory_store::<AB>(builder, local);
-
+*/
```
felicityin: It will cause the addition test to fail. Please run test `RUST_LOG=debug cargo test -r test_add_prove --features debug`

weilzkm: OK. I'll restore it.

eigmax (approved):

felicityin (approved):

eigmax (approved):

PR #60, opened 2025-02-12
feat: change mips support to Little endian

weilzkm:
use new version toolchain: https://github.com/zkMIPS/toolchain/releases/tag/just_for_test


crates/core/machine/src/syscall/precompiles/weierstrass/weierstrass_decompress.rs
```diff
@@ -562,7 +562,7 @@ mod tests {
         let mut rand = RAND::new();
 //
         let len = 100;
-        let num_tests = 10;
+        let num_tests = 2;
```
eigmax: plz recover this test number

felicityin (approved):

crates/core/machine/src/syscall/precompiles/weierstrass/weierstrass_decompress.rs
weilzkm: Ok

blake-pro (approved):

Issue #57, opened 2025-02-11
Add byte related opcode constraints


eigmax:
The opcode not constrained includes: LWR, LWL, LL, SWL, SWR, SC, SDC1.

position:  https://github.com/zkMIPS/zkm2/blob/dev/init/crates/core/machine/src/cpu/air/memory.rs#L253

VanhGer:
## Load memory opcode:

Corresponding variants between executor and memory trace:

| Executor | Memory                                            |
| -------- | ------------------------------------------------- |
| rt       | prev_a_val                                        |
| out      | a_val                                             |
| mem      | mem_val (value at Mem[virt])                      |
| rs & 3   | offset_is_one<br>offset_is_two<br>offset_is_three |
| virt     | addr_aligned                                      |

Note: which Load memory opcode, the mem_val and prev_mem_val are the same.

### LWR:

```rust
// Compute the output
let out = |i: u32| -> u32 {  
    let val = mem >> (i * 8);  
    let mask = 0xffFFffFFu32 >> (i * 8);  
    (rt & (!mask)) | val  
};  
out(rs & 3)
...

// Compute unsigned_mem_val
let val = mem_value >> (addr_offset * 8);  
let mask = 0xffFFffFFu32 >> (addr_offset * 8);  
cols.unsigned_mem_val = ((mem_value & (!mask)) | val).into();
```

We compute the `lwr_expected_load_value` value in Word representation like this:

| offset  = 0 | offset = 1 | offset = 2 | offset = 3 |
| ----------- | ---------- | ---------- | ---------- |
| mem[0]      | mem[1]     | mem[2]     | mem[3]     |
| mem[1]      | mem[2]     | mem[3]     | rt[1]      |
| mem[2]      | mem[3]     | rt[2]      | rt[2]      |
| mem[3]      | rt[3]      | rt[3]      | rt[3]      |

The `lwr_unsigned_mem_val` is computed by the same way

Constraints:
```rust
builder  
    .when(local.selectors.is_lwr)  
    .assert_word_eq(a_val.map(|x| x.into()), lwr_expected_load_value);

builder  
    .when(local.selectors.is_lwr)  
    .assert_word_eq(lwr_unsigned_mem_val, local.unsigned_mem_val.map(|x| x.into()));

```

### LWL

It's quite similar to LWR opcode:
``` rust
// Compute the output
let out = |i: u32| -> u32 {  
    let val = mem << (24 - i * 8);  
[...]

PR #64, opened 2025-02-14
feat: add byte related opcode constraints

VanhGer:
Implement issue #57


crates/core/machine/src/cpu/air/memory.rs
```diff
@@ -356,6 +491,58 @@ impl CpuChip {
 
         // When the instruction is LW, just use the word.
         builder.when(local.selectors.is_lw).assert_word_eq(mem_val, local.unsigned_mem_val);
+
+        // When the instruction is LWR:
+        //     val = mem_value >> (24 - addr_offset * 8);
+        //     mask = 0xffFFffFFu32 >> (24 - addr_offset * 8);
+        //     unsigned_mem_val = ((mem_value & (!mask)) | val);
+        let lwr_unsigned_mem_val = Word([
```
weilzkm, line 499: In little endian memory.  Opcode::LWR => {
                let out = |i: u32| -> u32 {
                    let val = mem >> (i * 8);
                    let mask = 0xffFFffFFu32 >> (i * 8);
                    (rt & (!mask)) | val
                };
                out(rs & 3)
            }

crates/core/executor/src/executor.rs
```diff
@@ -2721,4 +2718,4 @@ mod tests {
     //     assert_eq!(runtime.register(Register::X12), 0x12346525);
     //     assert_eq!(runtime.register(Register::X11), 0x65256525);
     // }
-}
+}
```
weilzkm: Remain an empty line at end

crates/core/executor/src/programs.rs
```diff
@@ -135,4 +135,30 @@ pub mod tests {
...
+
+    pub fn other_memory_program() -> Program {
+        let instructions = vec![
+            Instruction::new(Opcode::ADD, 29, 0, (1 << 20) + (1 << 15) + (1 << 6) - 1, false, true),
+            Instruction::new(Opcode::ADD, 27, 0, 25, false, true),
+            Instruction::new(Opcode::ADD, 25, 0, (1 << 28) + (1 << 12) + (1 << 18) - 1, false, true),
+            Instruction::new(Opcode::ADD, 17, 0, 0x43627530, false, true),
+            Instruction::new(Opcode::ADD, 22, 0, 22, false, true),
+            Instruction::new(Opcode::ADD, 10, 0, 15, false, true),
+            Instruction::new(Opcode::LWR, 29, 27, 1, false, true),
```
weilzkm, line 147: Normally, to access a memory address, we need do a  store before a load.

crates/core/machine/src/cpu/air/memory.rs
VanhGer: Yes. But this code computes the unsigned_mem_val, following this: 
https://github.com/zkMIPS/zkm2/blob/509d482d63c620c12e52d72ce59e0b9c9cc0b18a/crates/core/machine/src/cpu/trace.rs#L343

weilzkm: Yeah. It's wrong here, too.

eigmax (approved):

PR #86, opened 2025-03-19
feat: add more exampels

felicityin:


examples/is-prime/host/proof-with-is-prime.bin
eigmax, line 1: plz remove this file?

felicityin: ok

PR #97, opened 2025-03-24
style: refactor arkworks converter

felicityin:


crates/verifier/src/groth16/ark_converter.rs
```diff
@@ -35,6 +43,50 @@ pub enum ArkGroth16Error {
...
+    // the given groth16_vk.
+    //
+    // ZKM2 prepends the raw Groth16 proof with the first 4 bytes of the groth16 vkey to
+    // facilitate this check.
+    if groth16_vk_hash != proof[..4] {
+        return Err(ArkGroth16Error::Groth16VkeyHashMismatch);
+    }
+
+    // Convert gnark proof to arkworks proof
+    let ark_proof = load_ark_proof_from_bytes(&proof[4..]).unwrap();
```
eigmax: can we use modifier ?  instead of unwrap to handle the error?

eigmax (approved):

PR #104, opened 2025-03-28
opt: reduce revme examples cycles by switching io

felicityin:
Reduce the cycle count in the revme example from172,276 to 143,877 by switching io.


examples/revme/guest/src/main.rs
```diff
@@ -8,6 +8,6 @@ use guest::verify_revm_tx;
 zkm2_zkvm::entrypoint!(main);
 
 pub fn main() {
-    let input: Vec<u8> = zkm2_zkvm::io::read();
```
eigmax, line 11: what's the difference between read and read_vec?

felicityin: `zkm2_zkvm::io::read::<T>` is for reading structured data. The type `T` must implement both `serde::Serialize` and `serde::Deserialize`. 
For direct byte-level operations, use `zkm2_zkvm::io::read_vec`. It does not perform serialization and is substantially faster for reading in bytes directly.

eigmax (approved):

PR #108, opened 2025-03-31
feat: Keccak sponge precompile

VanhGer:
Design and Benchmark:

https://www.notion.so/Third-approach-1c8f0a489b98807fb4bdc8f648622725


crates/core/executor/src/syscalls/code.rs
```diff
@@ -52,8 +52,8 @@ pub enum SyscallCode {
     /// Executes the `ED_DECOMPRESS` precompile.
     ED_DECOMPRESS = 0x00_00_01_08,
 
-    /// Executes the `KECCAK_PERMUTE` precompile.
-    KECCAK_PERMUTE = 0x00_01_01_09,
```
eigmax, line 56: can we reuse the syscall KECCAK_PERMUTE no.?

VanhGer: Yes, I will do this

eigmax (approved):

felicityin (approved):

PR #135, opened 2025-04-12
doc syntax

lin677622:
doc syntax fix


docs/src/design/chips/other-components.md
```diff
@@ -34,7 +34,7 @@ zkMIPS's implementation integrates specialized permutation logic with KoalaBear
 
 Three proofs are used in zkMIPS
 - Shard Proofs: Used to verify correct execution of patched MIPS instructions (i.e., shard).
-- STARK Compressed Proof: Compress segements proof into one STARK proof.
+- STARK Compressed Proof: Compress segments proof into one STARK proof.
```
eigmax: Segment should be changed to Shard

lin677622:
segments change to shards

docs/src/design/chips/cpu.md
fulian89, line 1: CPU Chip is treated as a proper noun, so the first letter of 'Chip' is capitalized, and the same rule applies to the other instances of 'chip'. It's fine to replace all 'Chip' with 'chip' to keep consistent.

lin677622:
replace all 'Chip' with 'chip'

eigmax (approved):

PR #141, opened 2025-04-14
docs: intro and proof network

eigmax:


docs/src/dev/prover.md
```diff
@@ -138,3 +138,144 @@ To activate **AVX512** optimization, add these flags to your RUSTFLAGS environme
...
+}
+```
+
+
+> [!NOTE] 
+> The proof network uses `stdin.write_vec()` to write private input data to the guest program. 
+> If your guest program uses `zkm_zkvm::io::read();` to read this input, you must serialize 
+> it before pushing to the private input:
+```rust
+fn write_to_guest_private_input(private_input: &mut Vec<Vec<u8>>, data: &[u8]) {
```
eigmax, line 285: @VanhGer why should we need this?  no code calls this function.

VanhGer: Oh, I missed the code in the example. I will add it.

VanhGer: @eigmax I added the missing code:
https://github.com/zkMIPS/zkMIPS/pull/147

PR #155, opened 2025-04-16
doc: some updates

felicityin:
Solve this issue: https://github.com/zkMIPS/zkMIPS/issues/154


eigmax:
We need to update all R0's performance data. And we also need to introduce the version of all the other zkVMs in the doc

PR #177, opened 2025-05-06
feat: impl hooks for reth

felicityin:
1. Add hooks
   - FD_FP_SQRT: Compute the square root of a field element. Used by [k256](https://github.com/zkMIPS-patches/elliptic-curves/tree/patch-k256-0.13.4) and [p256](https://github.com/zkMIPS-patches/elliptic-curves/tree/patch-p256-0.13.2).
   - FD_FP_INV: Compute the inverse of a field element. Used by [k256](https://github.com/zkMIPS-patches/elliptic-curves/tree/patch-k256-0.13.4) and [p256](https://github.com/zkMIPS-patches/elliptic-curves/tree/patch-p256-0.13.2).

2. Fix keccak rows per event.
3. Fix bn254 multiplication.
4. Implement prover trait for ProverClient.
5. Regenerate shapes.


examples/unconstrained/guest/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+
+zkm_zkvm::entrypoint!(main);
+
+pub fn main() {
+    const MODULUS: &str = "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141";
+    const NQR: FieldElement = FieldElement::from_u64(3);
+    let value = FieldElement::from_u64(1);
+    let (status, result) =
+        call_sqrt_hook(value.to_bytes().as_slice(), MODULUS, NQR.to_bytes().as_slice());
+    println!("status: {:?}", status);
```
eigmax, line 13: can we add an assert here to check the result is as expected?

crates/core/machine/src/mips/mod.rs
```diff
@@ -563,14 +551,51 @@ impl<F: PrimeField32> MipsAir<F> {
...
+                0
+            }
             _ => 1,
         }
     }
 
+    fn keccak_rows_per_event(&self, record: &ExecutionRecord) -> usize {
+        let all = self.keccak_rows_per_record(record);
+
+        let count = record
```
VanhGer, line 575: can we return 0 when `all` is not > 0.  So we dont need to compute the `count`.

examples/unconstrained/guest/src/main.rs
felicityin: Actually, I don't know what the correct result should be. Currently, this example does not run successfully.

felicityin: This example is used to debug `zkm_lib::unconstrained!`, as the name of this use case suggests.

crates/core/machine/src/mips/mod.rs
felicityin: ok

examples/unconstrained/guest/src/main.rs
eigmax: Okay.  You can use sage to calculate it.  Or you can get the answer and then calculate the square, and verify  the result.  it's optional.

felicityin: ok

eigmax (approved):

PR #179, opened 2025-05-07
opt: reduce instructionCols columns

weilzkm:


crates/core/machine/src/cpu/air/register.rs
```diff
@@ -58,7 +58,7 @@ impl CpuChip {
         builder.eval_memory_access(
             local.shard,
             clk.clone() + AB::F::from_canonical_u32(MemoryAccessPosition::HI as u32),
-            local.instruction.op_hi[0],
+            AB::F::from_canonical_u32(33),
```
eigmax, line 61: Can we add more comments on this change?

PR #187, opened 2025-05-12
fix: support compressed proof

blake-pro:


crates/sdk/src/network/prover.rs
```diff
@@ -120,14 +120,25 @@ impl NetworkProver {
             .expect("connect: {self.endpoint:?}")
     }
 
-    async fn request_proof(&self, input: &ProverInput) -> Result<String> {
+    async fn request_proof(&self, input: &ProverInput, kind: ZKMProofKind) -> Result<String> {
+        let seg_size =
```
felicityin, line 124: Is a default value of 0 the expected behavior?

crates/sdk/src/network/prover.rs
```diff
@@ -240,9 +252,9 @@ impl Prover<DefaultProverComponents> for NetworkProver {
         stdin: ZKMStdin,
         _opts: ProofOpts,
         _context: ZKMContext<'a>,
-        _kind: ZKMProofKind,
+        kind: ZKMProofKind,
```
felicityin, line 255: Please update the description of this function.

crates/sdk/src/network/prover.rs
blake-pro: yes

felicityin (approved):

PR #197, opened 2025-05-25
Opt check for access to a and hi register

weilzkm:
- Delete unused op_a_0 from chips for instructions
- Instead of check in CPU chip，Check access for hi register in each chip for related instructions 
- Check consistency of prev value of a between instruction chips  and CpuChip by reuse field for hi value


examples/plonk/host/src/main.rs
```diff
@@ -25,10 +25,10 @@ fn generate_fibonacci_proof() -> (Vec<u8>, Vec<u8>, String) {
     // Create a `ProverClient`.
     let client = ProverClient::new();
 
-    // Generate the groth16 proof for the Fibonacci program.
```
eigmax, line 28: Should we change this? we can keep using the plonk, and fix the run_plonk.sh

weilzkm: This example program is for plonk. groth16 is another example program.

PR #198, opened 2025-05-28
Add missed constraints for reduced value，ins/ext instructions，and initial next pc

weilzkm:


crates/core/machine/src/control_flow/branch/air.rs
```diff
@@ -101,8 +101,8 @@ where
 
             KoalaBearWordRangeChecker::<AB::F>::range_check(
                 builder,
-                local.target_pc,
```
eigmax, line 104: should we check the range of target_pc?

eigmax (approved):

crates/core/machine/src/control_flow/branch/trace.rs
```diff
@@ -116,7 +116,7 @@ impl BranchChip {
         cols.target_pc = Word::from(target_pc);
         cols.next_next_pc = Word::from(event.next_next_pc);
         cols.next_pc_range_checker.populate(event.next_pc);
-        cols.target_pc_range_checker.populate(target_pc);
+        cols.next_next_pc_range_checker.populate(event.next_next_pc);
```
eigmax, line 119: Is it necessary to range check the target_pc?

weilzkm: next_next_pc will be target pc when branching.  target pc will be unused when not branching. So I think check on next_next_pc can replace  check on target pc.

PR #252, opened 2025-07-23
fix: cert is not required to exist

felicityin:
Fix cert path.


VanhGer:
Please change the document to adapt to this change.

VanhGer (approved):

PR #262, opened 2025-08-06
feat: add no-std stark verifier

felicityin:
Add no-std stark verifier.


crates/verifier/src/stark/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+const COMPRESS_DEGREE: usize = 3;
+pub type CompressAir<F> = RecursionAir<F, COMPRESS_DEGREE>;
+type CompressProver = CpuProver<InnerSC, CompressAir<<InnerSC as StarkGenericConfig>::Val>>;
+
+pub(crate) fn verify_stark_compressed_proof(
+    vk: &ZKMVerifyingKey,
+    proof: &ZKMReduceProof<InnerSC>,
+) -> Result<(), MachineVerificationError<InnerSC>> {
+    let allowed_vk_map: BTreeMap<[KoalaBear; DIGEST_SIZE], usize> =
+        bincode::deserialize(include_bytes!("../../../prover/dummy_vk_map.bin")).unwrap();
```
eigmax: It's better to use absolute path, and define it as a global variable,  like

const VK_MAP: &[u8] = include_bytes!(concat!(
    env!("CARGO_MANIFEST_DIR"), 
    "/prover/dummy_vk_map.bin"
));

We also need to make sure the other places that uses similar relative path to this way.  It's best to define a unified global variable for this.

felicityin: ok

eigmax (approved):

blake-pro (approved):

PR #277, opened 2025-08-15
repr usize

ple1n:
image

I was using zkm-verifier in WASM environment

The code didn't compile without the fix


crates/core/executor/src/syscalls/code.rs
```diff
@@ -24,6 +24,7 @@ use strum_macros::EnumIter;
 #[allow(clippy::enum_clike_unportable_variant)]
 #[allow(non_camel_case_types)]
 #[allow(clippy::upper_case_acronyms)]
+#[repr(usize)]
```
eigmax: By design, it's should be u32

ple1n: fixed. and compiles on wasm now

PR #280, opened 2025-08-18
feat: add new function `verify_proof()` for stark verifier.

felicityin:
1. Add new function `verify_proof()` for stark verifier.
2. Update allowed vk maps, groth16 vk and plonk vk.


crates/core/machine/src/lib.rs
```diff
@@ -39,7 +39,7 @@ pub use mips::*;
 /// This string should be updated whenever any step in verifying a Ziren proof changes, including
 /// core, recursion, and plonk-bn254. This string is used to download Ziren artifacts and the gnark
 /// docker image.
-pub const ZKM_CIRCUIT_VERSION: &str = "v1.1.2";
+pub const ZKM_CIRCUIT_VERSION: &str = "v1.1.3";
```
eigmax: Can we just detect the version in root cargo.toml?

crates/verifier/src/stark/verify.rs
```diff
@@ -33,19 +33,27 @@ lazy_static::lazy_static! {
...
     // ```
     // cd Ziren
-    // cargo run -r --bin build_compress_vks -- --reduce-batch-size 8 --num-compiler-workers 8 --count-setup-workers 8 --build-dir crates/prover
+    // cargo run -r --bin build_compress_vks -- --num-compiler-workers 32 --count-setup-workers 32 --build-dir crates/prover
     // ```
     // It takes several days.
     static ref VK_MAP: &'static [u8] =
         include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/../prover/vk_map.bin"));
+
+    static ref DUMMY_VK_MAP: &'static [u8] =
```
eigmax: can we use features to avoid loading 2 vk_map?  this can help reduce the memory usage

eigmax: and the code would be clear, since we have hardcoded the VERIFY_VK

crates/verifier/src/stark/mod.rs
```diff
@@ -23,6 +25,47 @@ use verify::verify_stark_compressed_proof;
 pub mod error;
 mod verify;
 
+/// A proof generated with Ziren of a particular proof mode.
+#[derive(Debug, Clone, Serialize, Deserialize, EnumDiscriminants, EnumTryAs)]
+#[strum_discriminants(derive(Default, Hash, PartialOrd, Ord))]
+#[strum_discriminants(name(ZKMProofKind))]
+pub enum ZKMProof {
```
eigmax, line 33: Do we really need this wrapping? we only use it in stark proof?

eigmax:
Plz remove https://github.com/ProjectZKM/Ziren/blob/main/crates/prover/scripts/e2e.rs.

crates/verifier/src/stark/verify.rs
felicityin: Because `const VERIFY_VK: bool = true;` and `lazy_static`, the `dummy_vk_map.bin` file will not be loaded.

felicityin: Ok, I will use features.

crates/verifier/src/stark/mod.rs
felicityin: Yes, it is required; otherwise, the `sdk` crate would need to be imported, which depends on `std`.

felicityin:
> Plz remove https://github.com/ProjectZKM/Ziren/blob/main/crates/prover/scripts/e2e.rs.

ok

crates/core/machine/src/lib.rs
felicityin: yes

eigmax (approved):

crates/verifier/src/stark/mod.rs
```diff
@@ -45,8 +89,30 @@ pub trait HashableKey {
...
+    /// # Returns
+    ///
+    /// A success [`Result`] if verification succeeds, or a [`StarkError`] if verification fails.
+    ///
+    /// Compared to `verify_proof()`, it performs a consistency check between
+    /// user-supplied public values and those committed in the proof.
     pub fn verify(proof: &[u8], zkm_public_inputs: &[u8], zkm_vk: &[u8]) -> Result<(), StarkError> {
-        let proof: ZKMReduceProof<InnerSC> = bincode::deserialize(proof).unwrap();
+        let proof: ZKMProof = bincode::deserialize(proof).unwrap();
+        let ZKMProof::Compressed(proof) = proof else { panic!() };
```
VanhGer: Should we add a panic message (e.g: "expected a Compressed proof") or return an error here?

VanhGer (approved):

PR #282, opened 2025-08-19
feat: add single node for network prover

VanhGer:


eigmax (approved):

crates/sdk/src/network/prover.rs
```diff
@@ -143,6 +143,10 @@ impl NetworkProver {
         let max_prover_num =
             env::var("MAX_PROVER_NUM").ok().and_then(|s| s.parse::<u32>().ok()).unwrap_or(0);
 
+        // set the single node mode
+        let single_node =
```
eigmax, line 149: Is this only useful in proof network?

eigmax: If so, plz add some comments for this

VanhGer: Got it

PR #293, opened 2025-08-29
Remove unused std::vec import in RSA example

aso20455:
Description: Remove unused `use std::vec;` import from the RSA host example. The Vec type and vec! macro are available in the prelude, making this import unnecessary and causing compiler warnings.


eigmax:
Can you plz correct the typo here?  [compte_vkey_digest_koalabear](https://github.com/ProjectZKM/Ziren/blob/main/crates/sdk/src/utils.rs#L72)

aso20455:
Hey @eigmax, ready, thank you.

PR #298, opened 2025-09-04
docs: fix broken link

radik878:
https://github.com/ProjectZKM/Ziren/blob/main/examples/fibonacci_c_lib/guest/src/c_lib/add.cpp - correct
https://github.com/ProjectZKM/Ziren/blob/main/examples/fibonacci_c_lib/lib/add.cpp - broken


eigmax:
it's not, you can try to access this https://docs.zkm.io/dev/guest-program.html

PR #319, opened 2025-09-29
feat: adapt to gpu

felicityin:
1. Generate cpp version of chip traces.
2. Regenerate vk_map.bin, groth16 vk, and plonk vk.


crates/recursion/core/src/chips/batch_fri.rs
```diff
@@ -72,49 +73,66 @@ impl<F: PrimeField32, const DEGREE: usize> MachineAir<F> for BatchFRIChip<DEGREE
...
-                rows.extend(row_add);
             });
+            rows.extend(row_add);
+        });
 
         // Pad the trace to a power of two.
         pad_rows_fixed(
             &mut rows,
-            || [F::ZERO; NUM_BATCH_FRI_PREPROCESSED_COLS],
+            || [KoalaBear::ZERO; NUM_BATCH_FRI_PREPROCESSED_COLS],
```
eigmax, line 119: Can we stick to use F::ZERO here?

crates/recursion/core/src/chips/fri_fold.rs
```diff
@@ -167,55 +145,71 @@ impl<F: PrimeField32, const DEGREE: usize> MachineAir<F> for FriFoldChip<DEGREE>
...
+        let events = unsafe {
+            std::mem::transmute::<&Vec<FriFoldEvent<F>>, &Vec<FriFoldEvent<KoalaBear>>>(
+                &input.fri_fold_events,
+            )
+        };
 
+        let mut rows = events
+            .iter()
+            .map(|event| {
+                let mut row = [KoalaBear::ZERO; NUM_FRI_FOLD_COLS];
```
eigmax, line 190: Same as previous comment, do we have to change F::ZERO to Koalabear::ZERO?

crates/recursion/core/src/chips/batch_fri.rs
felicityin: No, it will cause a compilation failure.

crates/recursion/core/src/chips/fri_fold.rs
felicityin: No, it will cause a compilation failure.

eigmax (approved):

PR #320, opened 2025-10-05
Add check for next_next_pc

weilzkm:
TODO： update vk and vk map


crates/stark/src/lookup/builder.rs
```diff
@@ -244,15 +244,15 @@ mod tests {
                 AirLookup::new(
                     vec![x.into(), y.into()],
                     AB::F::from_canonical_u32(3).into(),
-                    LookupKind::Alu,
+                    LookupKind::Byte,
```
eigmax, line 247: Why is it changed to Byte?

eigmax:
Plz update the version to 1.2.1

PR #334, opened 2025-10-25
feat: implement add_byte_lookup_events_from_maps for Vec<ByteLookupEv…

berZKerk:
## Summary
Implements the missing `add_byte_lookup_events_from_maps` method for `Vec<ByteLookupEvent>` in the `ByteRecord` trait implementation.

## Changes
- **File:** `crates/core/executor/src/events/byte.rs`
- **Lines:** 123-134
- **Change:** Replaced `todo!()` with proper implementation


eigmax:
Thanks. But there is no need to add byte event for byte chip itself.

PR #348, opened 2025-11-06
docs: typo in function name

hawkadrian:
There was a typo "posedion", changed it to "poseidon" in all the codebase,

It seems like there's no place left with this typo anymore, therefore code should be working as it worked before


eigmax:
Will delay to next version, cause it will affect the vk.

PR #373, opened 2025-12-04
feat: add check ciphertext precompile

VanhGer:
This precompile checks the correctness of ciphertext computations in AND and OR gates. Its main purpose is to accelerate the BitVM2-GC implementation: https://github.com/GOATNetwork/bitvm2-gc


crates/core/executor/src/air.rs
```diff
@@ -115,6 +115,8 @@ pub enum MipsAirId {
     SysLinux = 47,
     /// The MovCondChip.
     MovCond = 49,
+    // The CiphertextCheck chip.
+    CiphertextCheck = 50,
```
eigmax: The name is not making sense

crates/core/executor/src/record.rs
```diff
@@ -166,7 +167,27 @@ impl ExecutionRecord {
                     }
                     current_shard.push((syscall_event, event));
                 }
+                current_shard
+            } else if syscall_code == SyscallCode::CIPHERTEXT_CHECK {
+                let mut current_shard = Vec::new();
+                let mut current_len = 0;
 
+                for (syscall_event, event) in events {
+                    if let PrecompileEvent::CiphertextCheck(event) = &event {
```
eigmax: why not use match! here?  to reuse the code of branch PrecompileEvent::CiphertextCheck(event) = &event?

crates/core/machine/src/syscall/precompiles/ciphertext/air.rs
```diff
@@ -0,0 +1,231 @@
...
+use zkm_core_executor::syscalls::SyscallCode;
+use zkm_stark::{LookupScope, ZKMAirBuilder};
+
+impl<F> BaseAir<F> for CiphertextCheckChip {
+    fn width(&self) -> usize {
+        NUM_CIPHERTEXT_CHECK_COLS
+    }
+}
+
+impl<AB> Air<AB> for CiphertextCheckChip
```
eigmax: Add more description of this precompiles, like what the expression is.

crates/core/machine/src/syscall/precompiles/ciphertext/mod.rs
```diff
@@ -0,0 +1,12 @@
+mod air;
+mod columns;
+mod trace;
+
+#[derive(Default)]
+pub struct CiphertextCheckChip;
+
+impl CiphertextCheckChip {
```
eigmax: remove this, you can initialize with ::default()

crates/core/executor/src/air.rs
VanhGer: How about CiphertextVerify?

crates/core/executor/src/record.rs
VanhGer: Got it.

VanhGer: Because this code is under a different if-else branch to let PrecompileEvent::KeccakSponge(event) = &event

eigmax: can we do like:

match &event {
    PrecompileEvent::CiphertextCheck(ev) | PrecompileEvent::KeccakSponge(ev) => {
        println!("CiphertextCheck event: {:?}", ev);
    }
    
}

crates/core/executor/src/air.rs
eigmax: BoolearnCircuitGarbling?

crates/core/executor/src/record.rs
VanhGer: No, because the if let PrecompileEvent::CiphertextCheck(event) = &event should be used in case of syscall_code = SyscallCode::CIPHERTEXT_CHECK, and the PrecompileEvent::KeccakSponge(ev) should be put in case of SyscallCode::KECCAK_SPONGE

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/trace.rs
```diff
@@ -0,0 +1,225 @@
...
+                    unreachable!();
+                };
+
+                self.event_to_rows(event, &mut Vec::new())
+            })
+            .collect();
+
+        let num_real_rows = rows.len();
+        let padded_num_rows = num_real_rows.next_power_of_two();
+        for _ in num_real_rows..padded_num_rows {
```
eigmax: Use rows.resize or resize_with.

crates/core/executor/src/syscalls/precompiles/boolean_circuit/garble.rs
```diff
@@ -0,0 +1,111 @@
...
+        let delta: [u32; 4] = delta_u32s.try_into().unwrap();
+
+        // for each gate info
+        let mut gate_info_ptr = input_ptr + 20;
+        for _ in 0..num_gates_u32 {
+            let (gate_type_record, gate_type_u32) = ctx.mr(gate_info_ptr);
+            gate_read_records.push(gate_type_record);
+            gates_info.push(gate_type_u32);
+
+            let (h0_read_records, h0_u32s) = ctx.mr_slice(gate_info_ptr + 4, 4);
```
eigmax: We can read all the records and values with only one mr_slice call.

eigmax: mr_slice is expensive executions.

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/air.rs
```diff
@@ -0,0 +1,231 @@
...
+                local.shard,
+                local.clk,
+                local.input_address + AB::F::from_canonical_u32(4 + (i as u32) * 4),
+                &local.gates_input_mem[i + 1],
+                local.is_first_row,
+            );
+        }
+
+        // eval gate info read
+        for i in 0..17 {
```
eigmax: Plz change all the magic number to a global variable, and add comments.

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/air.rs
```diff
@@ -0,0 +1,231 @@
...
+        builder.assert_bool(local.is_and_gate);
+        builder.assert_bool(local.is_or_gate);
+        builder.assert_zero(local.is_and_gate * local.is_or_gate);
+
+        builder
+            .when(local.is_inner_row * local.is_and_gate)
+            .assert_eq(local.gate_type, AB::F::ZERO);
+        builder
+            .when(local.is_inner_row * local.is_or_gate)
+            .assert_eq(local.gate_type, AB::F::from_canonical_u32(7_u32));
```
eigmax: Magic number

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/columns.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub clk: T,
+    pub is_real: T,
+    pub input_address: T,
+    pub output_address: T,
+    pub receive_syscall: T,
+    pub is_first_row: T,
+    pub is_inner_row: T,
+    pub is_first_gate: T,
+    pub is_last_gate: T, // from first gate -> (last - 1)-th gate
+    pub not_last_gate: T,
```
eigmax: Can we use is_last_gate to represent not_last_gate (1 - ls_last_gate)?  For is_first_row, you can constrain with builder.when_first_row

VanhGer: we cannot represent not_last_gate by using 1 - is_last_gate, cause in the padding rows, the not_last_gate = is_last_gate = 0.

VanhGer: the builder.when_first_row or builder.is_first_row do not work inner `builder.eval_memory_access`

eigmax: > eval_memory_access

For padding row, is_real is false.

eigmax: > eval_memory_access

What's the reason for it's not working?

VanhGer: if I use: builder.when_first_row().eval_memory_access, it will over the constraint degree
If I use: builder.eval_memory_access(
     ....
     builder.is_first_row()
); 
it raise:
`
thread 'main' (168038) panicked at /home/vanhger/Work/zkMIPS/crates/stark/src/lookup/builder.rs:177:13:
not an affine expression in current row elements for first row
`

VanhGer: > > eval_memory_access
> 
> For padding row, is_real is false.

yes, but I use the not_last_gate to reduce the degree. If I replace it by is_real * (1 - is_last_gate) , the degree of this line will over 3: 
`builder
            .when(local.not_last_gate)
            .assert_eq(next.checks[3], local.checks[3] * next.checks[2]);
`

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+pub const GATE_INFO_BYTES: usize = 17;
+
+// OR gate id
+pub const OR_GATE_ID: u32 = 7;
+
+/// A chip that computes non-free-gate ciphertexts and verifies them against the received ones.
+pub struct BooleanCircuitGarbleChip;
+
+impl BooleanCircuitGarbleChip {
+    pub const fn new() -> Self {
```
eigmax: Derive it with Default

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/columns.rs
eigmax: > eval_memory_access
`we cannot represent not_last_gate by using 1 - is_last_gate, cause in the padding rows, the not_last_gate = is_last_gate = 0.`

How do you pad this table?  for padding row, the not_last_gate and ls_last_gate should be meaningless.

eigmax: > eval_memory_access

let is_first_row = builder.is_first_row();
eval_memory_access(,...., is_first_row)

VanhGer: > > eval_memory_access
> > `we cannot represent not_last_gate by using 1 - is_last_gate, cause in the padding rows, the not_last_gate = is_last_gate = 0.`
> 
> How do you pad this table? for padding row, the not_last_gate and ls_last_gate should be meaningless.

I pad with zeroes.  But the not_last_gate is not equal to 1 - is_last_gate in case of padding rows.

VanhGer: > > eval_memory_access
> 
> let is_first_row = builder.is_first_row(); eval_memory_access(,...., is_first_row)

I tried, it did not work.

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/air.rs
```diff
@@ -0,0 +1,231 @@
...
+        local: &BooleanCircuitGarbleCols<AB::Var>,
+    ) {
+        builder.assert_bool(local.is_real);
+        builder.assert_bool(local.is_first_gate);
+        builder.assert_bool(local.not_last_gate);
+        builder.assert_zero(local.is_last_gate * local.not_last_gate);
+        builder.assert_zero(local.is_last_gate * local.is_first_gate);
+        builder.assert_bool(local.gate_type[0]);
+        builder.assert_bool(local.gate_type[1]);
+        builder.assert_zero(local.gate_type[0] * local.gate_type[1]);
```
eigmax: this will be confusing, why not sum up all the gate type?

VanhGer: 2 gate types which are not boolean can have sum equal to 1 or 0 in finite field.

crates/core/machine/src/syscall/precompiles/boolean_circuit_garble/air.rs
```diff
@@ -55,12 +55,13 @@ impl BooleanCircuitGarbleChip {
         builder.assert_bool(local.is_real);
```
eigmax, line 55: plz check if we need this, cause we already constrain is_real in CpuChip, and all other chip should be enforced by that already, and propagate is_real by `send_instruction`.

eigmax: We also need to double check the other chips.

PR #375, opened 2025-12-10
fix: panic prone code

felicityin:


crates/core/executor/src/hook.rs
```diff
@@ -235,20 +241,32 @@ mod fp_ops {
...
+        if buf.len() < 4 {
+            return Err(ExecutionError::BufferLengthTooSmall(4, buf.len()));
+        }
+        let len_bytes: [u8; 4] = buf[0..4]
+            .try_into()
+            .map_err(|e: TryFromSliceError| ExecutionError::IntoArrayError(e.to_string()))?;
+        let len = u32::from_be_bytes(len_bytes) as usize;
+
+        if buf.len() != 4 + 2 * len {
+            tracing::error!("FpOp: Invalid buffer length");
```
eigmax, line 254: remove the tracing::error.

crates/curves/src/lib.rs
```diff
@@ -32,6 +32,47 @@ pub const COMPRESSED_POINT_BYTES: usize = 32;
...
+#[derive(Error, Debug, Serialize, Deserialize)]
+pub enum CurveError {
+    /// The execution failed while attempting to decompress a point on an elliptic curve from
+    /// its compact byte representation, indicating the provided data does not represent a valid point.
+    #[error("decompress point err, big-endian bytes: {:?}, sign: {1}", 0)]
+    InvalidAffinePoint(Vec<u8>, u32),
+
+    /// The execution failed while converting a slice to an array.
+    #[error("failed to convert slice {0} to array")]
+    IntoArrayError(String),
```
eigmax, line 45: Can we receive an Error here?  so we can construct IntoArrayError with any error instance.

crates/curves/src/weierstrass/secp256k1.rs
```diff
@@ -92,23 +92,38 @@ impl WeierstrassParameters for Secp256k1Parameters {
...
-pub fn secp256k1_decompress<E: EllipticCurve>(bytes_be: &[u8], sign: u32) -> AffinePoint<E> {
-    let computed_point =
-        k256::AffinePoint::decompress(bytes_be.into(), Choice::from(sign as u8)).unwrap();
+pub fn secp256k1_decompress<E: EllipticCurve>(
+    bytes_be: &[u8],
+    sign: u32,
+) -> Result<AffinePoint<E>, CurveError> {
+    let computed_point = k256::AffinePoint::decompress(bytes_be.into(), Choice::from(sign as u8))
+        .into_option()
+        .ok_or_else(|| CurveError::InvalidAffinePoint(bytes_be.to_vec(), sign))?;
```
eigmax, line 101: We try not to do any type conversion when raising an error.

crates/recursion/core/src/chips/mem/constant.rs
```diff
@@ -193,7 +204,7 @@ mod tests {
         };
         let chip = MemoryChip::default();
         let trace: RowMajorMatrix<KoalaBear> =
-            chip.generate_trace(&shard, &mut ExecutionRecord::default());
+            chip.generate_trace(&shard, &mut ExecutionRecord::default()).unwrap();
```
eigmax, line 207: propagate the exception.

crates/recursion/core/src/chips/mem/variable.rs
```diff
@@ -170,7 +181,7 @@ mod tests {
         };
         let chip = MemoryChip::default();
         let trace: RowMajorMatrix<KoalaBear> =
-            chip.generate_trace(&shard, &mut ExecutionRecord::default());
+            chip.generate_trace(&shard, &mut ExecutionRecord::default()).unwrap();
```
eigmax, line 184: propagate the error

felicityin: It's in a test.

crates/recursion/core/src/chips/mem/constant.rs
felicityin: It's in a test.

crates/core/executor/src/hook.rs
felicityin: This makes it easier to locate the error, since the same exception is thrown in several places.

crates/curves/src/lib.rs
felicityin: No, because `core::array::TryFromSliceError` does not implement the `Serialize` trait.

crates/curves/src/weierstrass/secp256k1.rs
felicityin: There doesn't seem to be a better way here.

crates/core/executor/src/hook.rs
eigmax: Define difference ExecutionError. The user can not access the proof network's log.

PR #411, opened 2026-01-06
chore(primitives): deprecate Buffer::head

SashaMalysehko:
`Buffer::head` is not used anywhere in the current codebase and is not part of any documented workflow. Marking it as deprecated keeps the public API technically intact while signalling that the method will be removed in a future release, reducing dead API surface without breaking existing users immediately.

### DIFF
--- a/crates/primitives/src/types.rs
+++ b/crates/primitives/src/types.rs
@@ -26,7 +26,10 @@ impl Buffer {
         Self { data: data.to_vec(), ptr: 0 }
     }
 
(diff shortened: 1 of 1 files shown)


eigmax:
I reverted this, since ZKMPublicValues is not exposing head function. 

However, it's necessary to expose head in ZKMPublicValues, for example, when we call read from ZKMPublicValues, we actually move the `ptr` to the head of the buffer, and then we can not use the buffer again.

PR #426, opened 2026-01-17
docs: align mov condition opcode names

kurahin:
Renamed MOVZ/MOVN to MEQ/MNE in the ISA table so the documentation matches the actual opcode names used by the executor and mnemonic mapping.


eigmax:
MNE is not the opcode name in MIPS, just an alias used inside Ziren.

PR #427, opened 2026-01-17
fix(executor): remove incorrect sign extension for J/JAL target address

reallesee:
J and JAL instructions use a 26-bit unsigned address field. The code was applying sign_extend<26> which corrupted jump targets with bit 25 set (addresses >= 0x08000000 after shift). Removed sign extension per MIPS spec.


eigmax:
Thanks for spotting out, can you help format the code?

reallesee:
Did it, thanks!

PR #428, opened 2026-01-17
fix(memory): use MAX_MEMORY in heap limit panic

yyhrnk:
The panic message hardcoded an outdated limit that conflicted with MAX_MEMORY and project docs. Using the constant in the formatted message keeps diagnostics accurate and prevents future drift if the limit changes.


eigmax:
Can you plz fix the lint?

yyhrnk:
> Can you plz fix the lint?

oh sorry, corrected

PR #437, opened 2026-01-26
fix: correct copy-paste errors in comments and imports

Fallengirl:
Found and fixed a few copy-paste mistakes where comments referenced wrong event types and constant names didn't match actual code. Also cleaned up a duplicate import that was commented out.


eigmax:
Thanks for the uptdate, can you rebase your branch and fix the clippy error?

Fallengirl:
@eigmax should be good now

PR #439, opened 2026-02-03
build: harden rerun-if-changed path handling

Olexandr88:
Removes unwrap from build script path handling and improves robustness of rerun-if-changed logic

### DIFF
--- a/crates/build/src/utils.rs
+++ b/crates/build/src/utils.rs
@@ -20,8 +20,14 @@ pub(crate) fn cargo_rerun_if_changed(metadata: &Metadata, program_dir: &Path) {
     ];
     for dir in dirs {
         if dir.exists() {
-            if let Ok(path) = dir.canonicalize() {
-                println!("cargo::rerun-if-changed={}", path.display());
+            if let Ok(canonical_path) = dir.canonicalize() {
+                println!("cargo::rerun-if-changed={}", canonical_path.display());
+            } else {
+                println!(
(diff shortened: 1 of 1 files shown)


crates/build/src/utils.rs
```diff
@@ -20,19 +20,21 @@ pub(crate) fn cargo_rerun_if_changed(metadata: &Metadata, program_dir: &Path) {
     ];
     for dir in dirs {
         if dir.exists() {
-            println!("cargo::rerun-if-changed={}", dir.canonicalize().unwrap().display());
+            if let Ok(path) = dir.canonicalize() {
+                println!("cargo::rerun-if-changed={}", path.display());
```
eigmax: cargo:rerun-if-changed is correct?

Olexandr88: This follows the modern Cargo build script directive (cargo::rerun-if-changed) and avoids unwrap() for robustness

felicityin (approved):

crates/build/src/utils.rs
felicityin:
```rust
if let Ok(canonical_path) = dir.canonicalize() {
    println!("cargo::rerun-if-changed={}", canonical_path.display());
} else {
    println!(
        "cargo::warning=Could not canonicalize path: {:?}, using original path",
        dir
    );
    println!("cargo::rerun-if-changed={}", dir.display());
}
```

felicityin (approved):

PR #442, opened 2026-02-09
Shankara/picus backend

shankarapailoor:
This PR adds a Picus backend for Ziren which allows developers to check the determinism of their circuits with Picus


crates/core/machine/src/alu/mul/mod.rs
```diff
@@ -112,15 +113,17 @@ pub struct MulCols<T> {
...
     /// Flag indicating whether the opcode is `MULT`.
+    #[picus(selector)]
     pub is_mult: T,
 
     /// Flag indicating whether the opcode is `MULTU`.
+    #[picus(selector)]
     pub is_multu: T,
 
-    /// Selector to know whether this row is enabled.
     pub is_real: T,
```
felicityin, line 127: Why remove the comments? Why isn't `# [picus (selector)]` added to this line?

shankarapailoor: For the `is_real` column we don't need a selector annotation - the extractor automatically sets `is_real = 1`. This was a comment I added before.

felicityin (approved):

crates/core/machine/src/control_flow/branch/air.rs
```diff
@@ -133,6 +133,11 @@ where
                 .when(local.is_branching)
                 .assert_word_eq(local.target_pc, local.next_next_pc);
 
+            builder.when(is_real.clone()).assert_eq(
```
eigmax: This is redundant.  
In the CPU chip, we already have:

 builder                                                                     
      .when_first_row()                                                       
      .when_not(local.is_halt)                                                
      .assert_eq(local.pc + AB::Expr::from_canonical_u32(4), local.next_pc); 

In current chip, local.next_pc is constrained by line 119 from the previous row.

shankarapailoor: thanks yeah i forgot to remove.

PR #448, opened 2026-02-14
fix: missing assert_bool for is_nor flag in BitwiseChip

eeemmmmmm:
Fixed copy-paste bug where is_xor was checked twice instead of checking is_nor. This left the is_nor boolean flag unconstrained in the AIR.


eigmax:
Thanks for spotting out, I will merge all the circuit changes to release-1.2.5.

PR #456, opened 2026-02-23
chore: avoid panic when reading cached r1cs metadata

Ocheretovich:
Handle metadata errors gracefully to avoid potential panic when checking cached r1cs files.


VanhGer (approved):
LGTM!

PR #463, opened 2026-03-02
fix(curves): correct nb_scalar_bits calculation in WeierstrassParameters

eeemmmmmm:
Fixed wrong multiplier in `nb_scalar_bits()` for Weierstrass curves.

Was: `NB_LIMBS * 16` = 512 bits
Now: `NB_LIMBS * NB_BITS_PER_LIMB` = 256 bits

This bug was inherited from SP1 (same issue exists in upstream). The code worked correctly but performed 2x unnecessary iterations in scalar multiplication (256 extra double operations with zero bits).

### DIFF
--- a/crates/curves/src/weierstrass/mod.rs
+++ b/crates/curves/src/weierstrass/mod.rs
@@ -43,7 +43,7 @@ pub trait WeierstrassParameters: EllipticCurveParameters {
     }
 
     fn nb_scalar_bits() -> usize {
-        Self::BaseField::NB_LIMBS * 16
+        Self::BaseField::NB_LIMBS * Self::BaseField::NB_BITS_PER_LIMB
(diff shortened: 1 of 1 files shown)


crates/curves/src/weierstrass/mod.rs
```diff
@@ -43,7 +43,7 @@ pub trait WeierstrassParameters: EllipticCurveParameters {
     }
 
     fn nb_scalar_bits() -> usize {
```
eigmax: Can we remove this overloading?

eeemmmmmm: Removed it. Now we just call `EllipticCurve::nb_scalar_bits()` directly in `sw_scalar_mul`.

PR #481, opened 2026-03-19
feat: build a groth16 vk that is not affected by Ziren upgrades

felicityin:
In order to avoid changes in groth16 vk after ziren upgrade, `vk_commitment` and `pc_start` will not be verified in the groth16 circuit, but will be placed in public inputs for verification.

When Ziren upgrades, only needs to update `cracks/verifier/bn254-vk/part_dark-vk.bin`, without updating `cracks/verifier/bn254-vk/imm_groth16_vk.bin`

Test
```sh
ZKM_IMM_WRAP_VK=1 cargo test -r --features ark -- --ignored test_verify_groth16_imm_wrap_vk
// or
cargo test -r --features imm-wrap-vk --features ark -- --ignored test_verify_groth16_imm_wrap_vk
```


crates/recursion/circuit/src/lib.rs
```diff
@@ -103,6 +103,14 @@ pub trait KoalaBearFriConfigVariable<C: CircuitConfig<F = KoalaBear>>:
         builder: &mut Builder<C>,
         public_values: RecursionPublicValues<Felt<C::F>>,
     );
+
+    fn commit_recursion_public_values_common(
+        _builder: &mut Builder<C>,
+        _public_values: RecursionPublicValues<Felt<C::F>>,
+        _vk_commitment: <Self as FieldHasherVariable<C>>::DigestVariable,
+        _pc_start: Felt<<C as Config>::F>,
+    ) {
```
eigmax: By default, we should check the public values in circuit.

crates/recursion/core/src/stark/utils.rs
```diff
@@ -12,3 +19,32 @@ pub fn zkm_dev_mode() -> bool {
     }
     enabled
 }
+
+/// Returns true if either the `ZKM_COMMON` environment variable is set or the `common` feature is enabled.
```
eigmax: Change ZKM_COMMON to ZKM_STARK_VERIFIER_WITHOUT_PUBLIC_INPUTS

crates/recursion/core/src/stark/utils.rs
```diff
@@ -12,3 +19,32 @@ pub fn zkm_dev_mode() -> bool {
...
+    let enabled = value == "1" || value.to_lowercase() == "true" || cfg!(feature = "common");
+    if enabled {
+        tracing::warn!("ZKM_COMMON environment variable is enabled.");
+    }
+    enabled
+}
+
+/// If zkm_common_mode() returns true, combine the base vkey hash with `vk_commitment` and `pc_start`
+/// using a Poseidon2 permutation.
+pub fn new_vk_hash(
```
eigmax: new_vk_commitment, differenciate from the real vk_hash.

crates/recursion/circuit/src/lib.rs
felicityin: Removed default implementation.

crates/recursion/core/src/stark/utils.rs
felicityin: Rename to ZKM_IMM_WRAP_VK.

crates/recursion/core/src/stark/utils.rs
felicityin: Rename to hash_vkey_with_part_vk.

PR #488, opened 2026-03-30
fix: syscall-air-soundness-bugs

eigmax:
This file tracks syscall-related AIR issues found during manual review and Picus analysis.

Findings: https://github.com/Veridise/Ziren/blob/shankara/syscall-bugs/crates/core/machine/src/syscall/SYSCALL_BUGS.md


felicityin (approved):

eigmax:
https://github.com/Veridise/Ziren/commit/776d688022dd2e076b84afd0fad74c6041b8fb26
The test constructs a shard whose visible global interaction multiset is:

• send(A)
• receive(B)
• with A != B.

That multiset is invalid and should not cancel under the intended semantics.

The PoC then forges only the Global chip witness columns by using the curve points for a different message C:

• row 0 witness point = send(C)
• row 1 witness point = receive(C)
• while leaving the tuple columns as send(A), receive(B).

Because the AIR checks only that the witness point is a valid signed curve point, but does not constrain point = Map(message, kind), the forged witness still satisfies the chip constraints. The accumulated global digest becomes zero and the proof verifies. Please take a look and let me know what you think.



The fix adds 7 constraints (lines 132-142) that link x_coordinate to        
  (values, kind, offset):                                                     
  - x[0] = values[0] + kind * 65536 — encodes the lookup kind in the upper    
  bits                                                                        
  - x[1..6] = values[1..6] — direct pass-through of message elements          
  - x[6] = values[6] * 256 + offset — encodes the lift_x offset in the low    
  bits                                                                        
                                                                              
  These constraints are guarded by when(is_real) so padding rows (which use   
  the dummy point) are unaffected.                                            
                                         
  Why this fixes the PoC: The attacker can no longer substitute a forged point
   for message C while keeping the tuple columns as A/B. The x-coordinate is
  now deterministically derived from the tuple columns, so point = Map(tuple) 
  is enforced. The only remaining degree of freedom is the y-coordinate, which
   is already constrained to be on the curve with the correct sign.

felicityin (approved):

eigmax:
### Column Impact (from `mips_costs.json`)

| Chip | Before | After | Delta | Description |
|------|--------|-------|-------|-------------|
| SyscallInstrs | 91 | 97 | +6 | `IsZeroOperation` for bidirectional `is_sys_linux` (#1) |
| SyscallCore | 6 | 11 | +5 | `arg1_lo/hi`, `arg2_lo/hi`, `result_lo/hi`, `is_linux`; removed redundant `arg1`/`arg2` (-2) |
| SyscallPrecompile | 6 | 11 | +5 | Same as SyscallCore |
| SysLinux | 44 | 81 | +37 | Bidirectional decoders (8 × IsZero), mmap nibbles (8 bits + IsZero), carry bits (2), composite flags (3), a0/a1 decoders (5 × IsZero) |
| GtColsBytes | 7 | 8 | +1 | `has_comparison` for complementary LTU lookup (#15) |

**Cost metric** (`cols × constraints`, from `mips_costs.json`):

| Chip | Before | After | Delta |
|------|--------|-------|-------|
| SyscallInstrs | 91 | 97 | +6 |
| SyscallCore | 39 | 41 | +2 |
| SyscallPrecompile | 39 | 41 | +2 |
| SysLinux | 187 | 187 | 0 (already accounted for in branch) |

### Interaction Impact

| Interaction | Change |
|-------------|--------|
| Global #1 (Syscall kind) | Message changed: `arg1/arg2` (reduced) → `arg1_lo/hi, arg2_lo/hi` (half-word packed). Same 7-element width, no new interaction. |
| Global #2 (SyscallResult kind) | **New**: `[shard, clk, syscall_id, result_lo, result_hi, 0, 0]`. +1 global interaction per syscall per shard. |
| U16Range byte lookups | +4 per SyscallChip row for half-word range checks. |
| Complementary LTU lookup | +1 per GtColsBytes usage (BRK comparison in SysLinux). |

### Optimizations Applied

| Optimization | Saving |
|-------------|--------|
| Removed `arg1`/`arg2` stored columns from SyscallChip | -2 columns per row. Derived inline as `arg1_lo + arg1_hi * 65536`. |
| Deduplicated `generate_dependencies` event iteration | 3 passes → 1 single pass over event triples. |
| Removed redundant `assert_eq` binding constraints | -2 degree-2 constraints (no longer needed since args are derived from half-words). |

### Overhead

Estimated **< 2% total proving overhead**. The SyscallChip is one of the narrowest and shortest chips (a few hundred rows per shard). The CPU (119), memory (100-127), and precompile chips (700-100k) dominate proving time with millions of rows.

PR #491, opened 2026-04-09
feat: trace FFI

felicityin:
Generate trace FFI for the following chips:
SyscallChip, LtChip, BitwiseChip, CloClzChip, BranchChip, JumpChip, MemoryInstrsChip, MiscInstrsChip, MovCondChip, ShiftLeftChip, ShiftRightChip, DivRemChip, MulChip, SyscallInstrsChip


crates/core/machine/include/mul.hpp
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include <vector>
```
eigmax: Unnecessary <vector>

crates/core/machine/include/utils.hpp
```diff
@@ -1,7 +1,10 @@
 #pragma once
 
+#include <vector>
```
eigmax: Unnecessary <vector>

felicityin: ok

crates/core/machine/include/mul.hpp
felicityin: ok

PR #493, opened 2026-04-10
ci: gpu 

felicityin:


.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - run: cargo clippy --all-targets -- -D warnings
   test:
     name: Cargo Test
-    runs-on: self-hosted
```
eigmax, line 45: all ci time-consuming task run in the new self-runner now?

PR #505, opened 2026-04-28
fix: bugs found by picus

felicityin:
Fixed bugs that were found by picus.
Basically, each commit corresponds to a fix for a counterexample.


felicityin:
Commit  [af1fe2a](https://github.com/ProjectZKM/Ziren/pull/505/commits/af1fe2a2cbc616e866d7b821d656fcbd716340d1)

Issue
- In the extracted Picus module, post-state memory limbs (state_mem_320, etc.) are not constrained to byte range, so the solver picks large field elements that still satisfy reconstruction equations.

Fix
- Add explicit <=255 constraints for the first three limbs of both pre-state and post-state words under
  is_real.

felicityin:
Strengthen constraints in Poseidon2PermuteChip​: [f67a502](https://github.com/ProjectZKM/Ziren/pull/505/commits/f67a50222803ad2db48f500919f88a8e64784af2)

- Issue: In the extracted Picus module, post-state memory limbs (state_mem_320, etc.) are not constrained to byte range, so the solver picks large field elements that still satisfy reconstruction equations.
- Fix: Add explicit <=255 constraints for the first three limbs of both pre-state and post-state words under
    is_real.

felicityin:
[82b96e1](https://github.com/ProjectZKM/Ziren/pull/505/commits/82b96e189efff435a36f84fa9ae6430c5d8280be): `ShaCompressCols` exports many `transition_output` fields (a..h, pointers, clocks) even though the AIR intentionally allows them to vary on the cycle’s terminal row (is_last_row = 1). I’m patching annotations to keep these as `transition_input` only, so Picus doesn’t demand false determinism on unconstrained successor state.

felicityin:
Strengthen constraints in KeccakSpongeChip​: [2173263](https://github.com/ProjectZKM/Ziren/pull/505/commits/2173263ae4eaa752f30549ee1b64ed6589fd6470)

The absence of a constraint linking `read_block` to `first_step` allows the case where `keccak_0 = 0` while `read_block = 1` in the model.

Patch is minimal and targeted:

  - Enforce `read_block == first_step * is_real` in eval_flags.
  - This makes block reads legal only on real rows at round 0, matching how trace rows are generated.

felicityin:
Strengthen constraints in KeccakSpongeChip​: [f8bedcc](https://github.com/ProjectZKM/Ziren/pull/505/commits/f8bedccc7f3acf724ef4236e37b1ad81b84276da)

What was wrong:

  - `input_address` had a jump constraint only on `is_absorbed` rows (next = local + 144), but no stability constraint on intermediate rounds.
  - That let the solver drift `input_address` across non-final rounds, which showed up as the `fresh_6493 / fresh_10024` mismatch class in the counterexample.

  What changed:

  - Added a non-final-round transition invariant:
    `when_transition && not_final_step => local.input_address == next.input_address`.
  - Kept the existing absorbed-step jump constraint, so behavior is now:
      - non-final round: address is constant
      - absorbed boundary step: address advances by one rate block (144 bytes)

felicityin:
Strengthen constraints in KeccakSpongeChip​: [87081b7](https://github.com/ProjectZKM/Ziren/pull/505/commits/87081b71ebc28f40518327b2c68333edf7a1b68e)

In `KeccakSpongeAir::eval_keccak_permutation`, when `read_block == 0`, constrain both `xored_general_rate[*].value[*]` and `block_mem[*].access.value[*]` bytes to zero, eliminating unconstrained byte-state drift observed in the counterexample.

felicityin:
Strengthen constraints in KeccakSpongeChip​: [a05197e](https://github.com/ProjectZKM/Ziren/pull/505/commits/a05197e229fc5228c29e44709db64f327d0c7cbe)

- Change: `cols.input_len` now uses `event.input_len_u32s` instead of `event.input.len() as u32`.
- Why: AIR constraints tie `input_len` semantics to the dedicated memory read (`input_length_record / input_length_mem`). Using the vector length introduced a second source of truth and could produce Picus counterexamples on final-block absorb/output constraints.
- Result: Trace semantics now match the AIR’s memory-backed input-length contract.

felicityin:
Strengthen constraints in KeccakSpongeChip​: [26e10e2](https://github.com/ProjectZKM/Ziren/pull/505/commits/26e10e2aac2eecd41469d6542ba31f2a7090529a)

What was wrong:
- `next.original_state is` consumed as packed bytes (`b0 + 256*b1`, etc.), but those bytes were not range-constrained. That allows invalid field elements to satisfy packing equalities and create bogus Keccak-state matches.

What changed:
- Added explicit U8Range constraints for all bytes of `local.original_state,` gated by `local.is_real`:
  - flatten 50 words × 4 bytes = 200 bytes
  - call `builder.slice_range_check_u8(&original_state_bytes, local.is_real)`

  This ensures every `original_state[i][j]` is a real byte before it is used in limb packing constraints (including those involving `next.original_state` in absorbed transitions).

felicityin:
Strengthen constraints in KeccakSpongeChip: [1934c91](https://github.com/ProjectZKM/Ziren/pull/505/commits/1934c91ed5dd9d11072a5952ee1d34ba0dc864d4)

- What was wrong

  `original_state` did not have an explicit transition-stability constraint on rows where the sponge is not crossing an absorbed edge.
  In practice, that left unnecessary witness freedom: for non-absorbed, non-final-output rows, `next.original_state` could drift without being tied back to `local.original_state`.

- What changed

  Added a transition guard in keccak_sponge/air.rs:

  - Guard: `(1 - is_absorbed) * (1 - write_output)` under `when_transition()`
  - Constraint: for all `i in 0..KECCAK_STATE_U32S`, enforce
    `local.original_state[i] == next.original_state[i]` (word-wise via `assert_word_eq`)

  This pins original_state across ordinary round transitions while intentionally excluding:

  - absorbed-edge transitions (`is_absorbed = 1`), where state handoff is constrained differently
  - final output rows (`write_output = 1`), whose successor is padding/non-semantic

felicityin:
Strengthen constraints in KeccakSpongeChip: [a93cac6](https://github.com/ProjectZKM/Ziren/pull/505/commits/a93cac6f7894890d74fa06877216658cb51b8d06)

- What was wrong
  `KeccakSpongeChip` was packing memory word limbs into larger values (`l0 + 256*l1`, then u16/u64 limb comparisons) without explicitly constraining those memory limbs to byte range.
  That left room for invalid witnesses (e.g. limb values > 255) to satisfy packed equalities, which is exactly what the counterexample was exploiting.

- What changed

    - Added explicit byte range checks in AIR for memory words that participate in packed-limb constraints:
      - `input_length_mem.value` (gated by `receive_syscall`)
      - `block_mem[*].access.value` (gated by `read_block`)
      - `output_mem[*].value` (gated by `write_output`)

  - Updated trace generation to emit matching byte lookups for those same memory values:

    - input read records
    - input-length record
    - output write records

    This aligns witness generation with the strengthened AIR and closes the unconstrained-byte gap on  memory-backed packed limbs.

felicityin:
Strengthen constraints in KeccakSpongeChip: [f108edd](https://github.com/ProjectZKM/Ziren/pull/505/commits/f108edd14ed668409db5f748efe8db5e01015494)

- What was wrong
  `KeccakSpongeChip` relied on the embedded Keccak sub-AIR to enforce step-flag semantics. Under the summarized boundary, we observed a counterexample where a single real row simultaneously satisfied `read_block = 1` and `write_output = 1`, which implies both `first_step = 1` and `final_step = 1`. That should be impossible for a valid Keccak round schedule, but it was not explicitly ruled out by the outer sponge AIR.

- What changed
    Added defensive outer-AIR constraints in eval_flags for the projected step flags:

    - Enforced booleanity of `first_step` and `final_step` (gated by `local.is_real`).
    - Enforced mutual exclusion `first_step * final_step == 0` (gated by `local.is_real`).

  This makes the sponge AIR robust even when Keccak internals are summarized, and blocks degenerate witnesses that collapse first/final round behavior into the same row.

felicityin:
Strengthen constraints in KeccakSpongeChip: [1542de6](https://github.com/ProjectZKM/Ziren/pull/505/commits/1542de6e0b2b99a85e34a313b3ca9039842cbc3b)

- What was wrong

    - `input_len` (scalar column) was not constrained to equal the 4-byte word read from `input_length_mem`.
    - So the prover could choose an arbitrary `input_len` (used in `already_absorbed_u32s = input_len - RATE`) while giving unrelated memory bytes.

- What changed

  In `crates/core/machine/src/syscall/precompiles/keccak_sponge/air.rs`, under the `input_length_mem` read, added: `when(receive_syscall): input_len == reduce(input_length_mem.value())`

felicityin:
Strengthen constraints in KeccakSpongeChip: [62a893b](https://github.com/ProjectZKM/Ziren/pull/505/commits/62a893b6a3b3eae73355f735971ce9e37d999868)

- What was wrong

    - `input_len` was only constrained against `input_length_mem` on `receive_syscall` rows.
    - On other real rows (especially `write_output` / final rows), `input_length_mem` could drift without being tied back to input_len.
    - The existing read-side `value == prev_value` check did not protect this, because for `MemoryReadCols`, `prev_value` aliases value (tautological).
- What changed
    - In `keccak_sponge/air.rs` (`eval_memory_access`), the binding:
      - `input_len == reduce(input_length_mem.value())`
      - was widened from `when(receive_syscall)` to `when(is_real)`.
    - This enforces the same `input_len/memory-word` consistency on every real row, closing the unconstrained-path used by the counterexample.

felicityin:
Commit: [8c7dc92](https://github.com/ProjectZKM/Ziren/pull/505/commits/8c7dc9267015e84500c55d47c1f060c7babdb4c3)

- What was wrong:

    - `KeccakSponge` exposed several columns as both Picus `transition_input` and `transition_output`.
    - On the last real row (`write_output = 1`), some of those “output” values are not required to match a meaningful next real state (next row is padding/non-real).
    - Picus then treated unconstrained next-row values as observable outputs, which produced counterexamples (fresh_6493, fresh_6502, etc.).

- What changed:

    - In crates/core/machine/src/syscall/precompiles/keccak_sponge/columns.rs, those unstable fields were changed from:
      - `#[picus(transition_input, transition_output)]`
      - to `#[picus(transition_input)]`

    - Fields changed:
      - shard, clk, input_address, output_address, input_len, already_absorbed_u32s, is_first_input_block, is_final_input_block, original_state.

crates/core/machine/src/alu/add_sub/mod.rs
```diff
@@ -85,7 +88,14 @@ impl<F: PrimeField32> MachineAir<F> for AddSubChip {
     }
 
     fn picus_info(&self) -> PicusInfo {
```
eigmax, line 91: can we add a enabling feature in this function?

crates/core/machine/src/cpu/air/mod.rs
```diff
@@ -169,7 +192,6 @@ impl CpuChip {
             .assert_eq(local.next_next_pc, next.next_pc);
 
         builder
-            .when_transition()
```
eigmax, line 172: for the 1st and last row, does this constraint satisfy?

eigmax:
For the last 4 or 5 comments, some are invalid, we should delete.

crates/core/machine/src/alu/add_sub/mod.rs
felicityin: Added
```
fn picus_info(&self) -> PicusInfo {
      #[cfg(feature = "picus")]
      {
          AddSubCols::<u8>::picus_info()
      }
      #[cfg(not(feature = "picus"))]
      {
          zkm_stark::PicusInfo::default()
      }
  }
```

crates/core/machine/src/cpu/air/mod.rs
felicityin: Yes.

felicityin:
> For the last 4 or 5 comments, some are invalid, we should delete.

I don't understand.

crates/core/machine/src/alu/add_sub/mod.rs
felicityin: ok

PR #522, opened 2026-08-01
feat: add blake3 option for second public input

VanhGer:


crates/primitives/src/io.rs
```diff
@@ -3,6 +3,20 @@ use num_bigint::BigUint;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
+/// Returns true if either the `ZKM_IMM_WRAP_VK` environment variable is set or the `imm-wrap-vk`
+/// feature is enabled.
+/// By default, the variable is disabled.
+pub fn zkm_imm_wrap_vk_mode() -> bool {
+    let value = std::env::var("ZKM_IMM_WRAP_VK").unwrap_or_else(|_| "false".to_string());
```
eigmax, line 10: do we have to use std in the zkm::primitives? any other way to distinguish whether ZKM_IMM_WRAP_VK is enabled?

VanhGer: zkm-primitives is already a plain std crate everywhere, not no_std.

VanhGer: right now it just has only way which checks the env ZKM_IMM_WRAP_VK is set to 1 or 0.

